PR #27, opened 2024-02-16
[TF FE] Del TF Translators

apaniukov:
Unblock layer tests in OpenVINO


rkazants:
@ilya-lavrenov, could you please merge this PR? It is needed for importing `openvino-tokenizers` in layers tests in OV main repo.
We are going to add layer tests for operations that are enabled via `openvino-tokenizers`.

ilya-lavrenov (approved):

PR #42, opened 2024-02-27
Add String Mapping Op

apaniukov:
**Ticket:** 132440


python/openvino_tokenizers/tokenizer_pipeline.py
```diff
@@ -274,10 +274,6 @@ def byte_level_splitter(cls) -> "RegexSplitStep":
             behaviour="isolate",
         )
 
-    @classmethod
-    def add_whitespace_to_the_next_word(cls):
-        return cls(r"\s\S", invert=False, behaviour="merge_with_next")
-
```
rkazants, line 280: do we need to remove it? please check

apaniukov: Yes, but not necessary in this PR.

src/vocab_encoder.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+
+VocabEncoder::VocabEncoder (
+    const ov::OutputVector& arguments,
+    std::shared_ptr<std::map<std::vector<uint8_t>, int>> vocab,
+    int default_value
+) :
+    ov::op::Op(arguments), m_vocab(vocab), m_default_value(default_value) {
+    if (m_vocab == nullptr) {
+        auto packed_vocab_const = as_type_ptr<Constant>(arguments[3].get_node_shared_ptr()->get_input_node_shared_ptr(0));
```
rkazants, line 25: please fix this to use unpacked string tensor format for vocab

rkazants: Also, this input may be not necessarily  of Constant type

src/vocab_encoder.cpp
```diff
@@ -0,0 +1,91 @@
...
+    FRONT_END_GENERAL_CHECK(this->get_input_element_type(6) == element::i32, "Expected an i32 tensor for VocabEncode values.");
+    // vocab.size == vocab_values.size when vocab is static
+    FRONT_END_GENERAL_CHECK(
+        this->get_input_partial_shape(3).is_dynamic() || this->get_input_partial_shape(3) == this->get_input_partial_shape(6),
+        "Expected equal number of vocab keys and values."
+    );
+    // default value
+    FRONT_END_GENERAL_CHECK(this->get_input_element_type(7) == element::i32, "Expected an i32 scalar for VocabEncode default value.");
+    // one data output, reuse ragged dimensions from split
+    this->set_output_type(0, element::i32, get_input_partial_shape(0));
```
rkazants, line 65: Please support `i64` type for values

rkazants (changes requested):
Now it is not suitable for TensorFlow needs:
1. format of vocal keys needs to be changed
2. need support of i64 values

rkazants (approved):

PR #56, opened 2024-03-05
Revert "Switch on TF FE extensions build in GHA"

mryzhov:
Reverts openvinotoolkit/openvino_tokenizers#51


ilya-lavrenov (changes requested):
Should not be merged

PR #123, opened 2024-04-29
[WHEEL] Freeze scikit-build-core version

mryzhov:
The new scikit-build-core version skips manylinux tag as one of the valid wheels tags:

https://github.com/scikit-build/scikit-build-core/pull/698


henryiii:
This was being incorrectly selected (due to what I believe is a bug in `packaging`). The wheels are not guaranteed to be `manylinux` wheels until you run them through `auditwheel`, which verifies they don't use anything that is not part of the manylinux spec, and bundles any dependencies not available in `manylinux` into the wheel with name mangling. Until that is run, the wheel is `linux`, which is not redistributable. `manylinux` ensures the user that this process has been run.

`cibuildwheel` does all this for you, or you can do it manually. If you really want to fake it, you can use `wheel tag` to change the tag manually.

henryiii:
The correct fix is to run auditwheel to produce the manylinux wheels, like this:

```console
$ auditwheel repair openvino_tokenizers-2024.1.0.0-81-py3-none-linux_x86_64.whl --exclude libopenvino_tensorflow_frontend.so.2410 --exclude libopenvino.so.2410
INFO:auditwheel.main_repair:Repairing openvino_tokenizers-2024.1.0.0-81-py3-none-linux_x86_64.whl
INFO:auditwheel.lddtree:Excluding libopenvino_tensorflow_frontend.so.2410
INFO:auditwheel.lddtree:Excluding libopenvino.so.2410
INFO:auditwheel.wheeltools:Previous filename tags: linux_x86_64
INFO:auditwheel.wheeltools:New filename tags: manylinux_2_17_x86_64, manylinux2014_x86_64
INFO:auditwheel.wheeltools:Previous WHEEL info tags: py3-none-linux_x86_64
INFO:auditwheel.wheeltools:New WHEEL info tags: py3-none-manylinux_2_17_x86_64, py3-none-manylinux2014_x86_64
INFO:auditwheel.main_repair:
Fixed-up wheel written to /wheelhouse/openvino_tokenizers-2024.1.0.0-81-py3-none-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
```

See https://github.com/scikit-build/scikit-build-core/pull/698#issuecomment-2085759622 for more info. This is more secure, as it protects you if any other things that are non-compliant sneak in over time. (You can also use cibuildwheel; with cibuildhweel, you can override the repair command to add the excludes above).

PR #125, opened 2024-04-29
Add Sequence Post-Processor Support

apaniukov:
- Support the new version of meta-llama3
- Add `not-add-special-tokens` flag during conversion
- Update fuzzing script


mryzhov (approved):

mryzhov:
do we need to merge it to release branches too?

apaniukov:
> do we need to merge it to release branches too?

We could port the first commit with the fix for llama3 tokenizer to 2024.1 release.

PR #128, opened 2024-05-03
[HF] Reorder tensorflow and sentencepiece linking

mryzhov:
Change the linking order based on the operating system to prevent segmentation faults caused by conflicts in protobuf versions


ilya-lavrenov (approved):
Ok to merge if it works in conda-forge

PR #148, opened 2024-05-17
Package into correct dirs

Wovchena:


Wovchena:
@apaniukov, @mryzhov, @mryzhov

apaniukov:
We don't need to apply this folder structure when building the wheel, that is why the tests failed.

src/CMakeLists.txt
Wovchena, line 1: > We don't need to apply this folder structure when building the wheel, that is why the tests failed.

Fixed. The tests still fail but they fail in empty PR the same way: https://github.com/openvinotoolkit/openvino_tokenizers/pull/149

apaniukov (approved):

apaniukov:
build_jenkins

akladiev:
build_jenkins

akladiev:
build_jenkins

PR #160, opened 2024-05-29
Switch default skip tokens flag behavior

slyalin:
Changes made in both CLI and API. API changes will directly affect how optimum-intel converts tokenizers as it doesn't override defaults. Keep old options for the CLI part for backward compatibility.

### DIFF
--- a/python/openvino_tokenizers/cli.py
+++ b/python/openvino_tokenizers/cli.py
@@ -58,24 +58,54 @@ def get_parser() -> ArgumentParser:
             "Example: `convert_tokenizer SimianLuo/LCM_Dreamshaper_v7 --subfolder tokenizer`"
         ),
     )
-    parser.add_argument(
+    special_tokens_group = parser.add_mutually_exclusive_group()
+    special_tokens_group.add_argument(
+        "--add-special-tokens",
+        "--add_special_tokens",
+        required=False,
+        action="store_true",
+        help=(
+            "Tokenizer will add special tokens during tokenization, similar to "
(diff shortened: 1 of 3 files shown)


python/openvino_tokenizers/cli.py
```diff
@@ -58,24 +58,54 @@ def get_parser() -> ArgumentParser:
...
         "--not-add-special-tokens",
         "--not_add_special_tokens",
         required=False,
         action="store_false",
+        default=False,
         help=(
             "Tokenizer won't add special tokens during tokenization, similar to "
-            "huggingface_tokenizer.encode(texts, add_special_tokens=False). Not affects tiktoken-base tokenizers."
+            "huggingface_tokenizer.encode(texts, add_special_tokens=False). Not affects tiktoken-base tokenizers. "
+            "This is the default behaviour, so adding this option has no effect, for backward compatibility only. "
```
ilya-lavrenov, line 82: Looks like it's not aligned with HF
image
why do we need to skip special tokens during tokenization? We have talked only about detokenization..

ilya-lavrenov:
Please, create PR to releases/2024/2

python/openvino_tokenizers/cli.py
slyalin: You wanted to add them manually as I remember. Too many variants are discussed without a clear summary: what do we want.

Wovchena:
build_jenkins

ilya-lavrenov (approved):

PR #173, opened 2024-06-11
Fix Tokenization of Special Tokens in Sentencepiece

apaniukov:
CVS-143410


python/openvino_tokenizers/hf_parser.py
```diff
@@ -333,7 +333,7 @@ def decoding(
 
 def parse_special_tokens(hf_tokenizer: PreTrainedTokenizerBase, only_special_tokens: bool = True) -> Dict[int, str]:
     # the order matters
-    if hasattr(hf_tokenizer, "added_tokens_decoder"):
+    if getattr(hf_tokenizer, "added_tokens_decoder", None):
```
rkazants, line 336: can we have tests in precommit for these cases to avoid regressions?

apaniukov: We [run tests](https://github.com/openvinotoolkit/openvino_tokenizers/actions/runs/9463408998/job/26069292457?pr=173) for [tokenizer](https://github.com/apaniukov/openvino_tokenizers/blob/3b0bd3b48c4836d358c1b665ad57ec7ba05c79f5/tests/tokenizers_test.py#L333)/[detokenizer](https://github.com/apaniukov/openvino_tokenizers/blob/3b0bd3b48c4836d358c1b665ad57ec7ba05c79f5/tests/tokenizers_test.py#L357) conversion that check regressions for a broad range of [sentencepiece models](https://github.com/apaniukov/openvino_tokenizers/blob/3b0bd3b48c4836d358c1b665ad57ec7ba05c79f5/tests/tokenizers_test.py#L119).

This line is to avoid the ChatGLM2 detokenizer regression shown by the tests.

PR #219, opened 2024-08-09
Ops Optimization

apaniukov:
Update VocabEncoder:
* Replace std::map with absl::flat_hash_map

Parallelize Operations:
* VocabEncoder
* EqualStr
* All Normalization Ops

Add batch_size argument to benchmark.py

CVS-148967
CVS-148968

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -211,7 +211,7 @@ def main(
         dump_latency_stats(result_df, model_name)
 
     title = (
-        f"OV vs HF Latency\n{checkpoint}\n"
+        f"OV vs HF Latency\n{checkpoint}, batch_size={batch}\n"
         f"OV: {ov_fps:.1f} FPS; "
         f"OV Async {async_fps:.1f} FPS; "
         f"HF  {hf_fps:.1f} FPS"
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -1,10 +1,10 @@
 import argparse
 import json
 import random
-from itertools import chain
+from itertools import chain, islice
 from random import sample, shuffle
 from time import perf_counter
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, Iterable
 
 import matplotlib.pyplot as plt
 import pandas as pd
@@ -32,7 +32,13 @@ def sample_texts(
     return dataset
 
 
-def benchmark_tokenizer_async(ov_tokenizer: CompiledModel, dataset: List[Tuple[str, str]]) -> Tuple[pd.Series, float]:
+def batch_iter(dataset: Iterable, batch: int = 1):
+    dataset_iter = iter(dataset)
+    while next_batch := list(islice(dataset_iter, batch)):
+        yield next_batch
+
+
+def benchmark_tokenizer_async(ov_tokenizer: CompiledModel, dataset: List[Tuple[str, str]], batch: int = 1) -> Tuple[pd.Series, float]:
     def callback(
         ir: InferRequest,
         user_data: Tuple[List[int], float, int],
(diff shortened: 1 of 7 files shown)


benchmark/benchmark.py
```diff
@@ -32,7 +32,13 @@ def sample_texts(
...
 
 
-def benchmark_tokenizer_async(ov_tokenizer: CompiledModel, dataset: List[Tuple[str, str]]) -> Tuple[pd.Series, float]:
+def batch_iter(dataset: Iterable, batch: int = 1):
+    dataset_iter = iter(dataset)
+    while next_batch := list(islice(dataset_iter, batch)):
+        yield next_batch
+
+
+def benchmark_tokenizer_async(ov_tokenizer: CompiledModel, dataset: List[Tuple[str, str]], batch: int = 1) -> Tuple[pd.Series, float]:
```
rkazants, line 41: looks unrelated changes to optimizations. May be it is better to do it in separate PR? Let us do it.

src/equal_str.cpp
```diff
@@ -41,28 +43,27 @@ bool EqualStr::evaluate(ov::TensorVector& outputs, const ov::TensorVector& input
...
         if ((end1 - begin1) != (end2 - begin2)) {
             // unequal string lengths guarantees false result
             result[idx] = 0;
-            continue;
         } else if (end1 == begin1) {
             // both empty strings case
             result[idx] = 1;
-            continue;
-        }
+        };
```
rkazants, line 61: let us remove this if-else construction:
1. auto chars1 = inputs[2].data<const char>();
2. std::string str1(chars1 + begin1, chars1 + end1);
3. result[idx] = (str1 == str2 ? 1 : 0);

src/vocab_encoder.cpp
```diff
@@ -61,14 +63,14 @@ bool VocabEncoder::evaluate(ov::TensorVector& outputs, const ov::TensorVector& i
...
 
-    for (size_t element_idx = 0; element_idx < num_elements; ++element_idx) {
-        auto element = m_vocab->find(std::vector<uint8_t>(chars + begins[element_idx], chars + ends[element_idx]));
+    ov::parallel_for(num_elements, [&](size_t element_idx){
+        const auto element = m_vocab->find(std::string(chars + begins[element_idx], chars + ends[element_idx]));
         if (element == m_vocab->end()) {
             token_ids[element_idx] = default_value;
         } else {
             token_ids[element_idx] = element->second;
         };
```
rkazants, line 72: please check if we can remove if-else and use default_value for all non-existent key.

src/equal_str.cpp
```diff
@@ -41,28 +43,19 @@ bool EqualStr::evaluate(ov::TensorVector& outputs, const ov::TensorVector& input
...
+        const size_t idx1 = (idx < num_elems1) ? idx : 0;
+        const size_t idx2 = (idx < num_elems2) ? idx : 0;
+        const auto begin1 = begins1[idx1];
+        const auto begin2 = begins2[idx2];
+        const auto end1 = ends1[idx1];
+        const auto end2 = ends2[idx2];
+
+        std::string op1(chars1 + begin1, chars1 + end1);
+        std::string op2(chars2 + begin2, chars2 + end2);
+        result[idx] = op1 == op2;
```
rkazants, line 57: for clarity
```suggestion
        result[idx] = (op1 == op2);
```

benchmark/benchmark.py
apaniukov: It is for better optimization testing for other tokenizers too.

src/vocab_encoder.cpp
```diff
@@ -45,11 +47,11 @@ bool VocabEncoder::evaluate(ov::TensorVector& outputs, const ov::TensorVector& i
         auto vocab_chars  = inputs[5].data<const uint8_t>();
 
         auto vocab_values = inputs[6].data<const int32_t>();
-        auto vocab_size = inputs[6].get_size();
+        const auto vocab_size = inputs[6].get_size();
 
-        m_vocab = std::make_shared<std::map<std::vector<unsigned char>, int32_t>>();
+        m_vocab = std::make_shared<absl::flat_hash_map<std::string, int32_t>>();
         for (size_t i = 0; i < vocab_size; ++i) {
```
rkazants, line 53: can we parallelize it?

rkazants (approved):
Please double-check that TF FE layer tests are passing

src/vocab_encoder.cpp
apaniukov: It could speed up the first inference, will do it later.

PR #231, opened 2024-08-23
Fix Segfault on Mac

pavel-esir:
`ovector` in PCRE2 wrapper was pointing to the freed memory location. Need to copy values to results and free only after that.


ilya-lavrenov (approved):
Please, port to 2024.3 if it's valid for that code base

pavel-esir:
> Please, port to 2024.3 if it's valid for that code base

For 2024.3 is not valid.

PR #268, opened 2024-10-02
Store tokenizer conversion params in rt_info / refactor passing params

pavel-esir:
CVS-146675

### DIFF
--- a/python/openvino_tokenizers/convert_tokenizer.py
+++ b/python/openvino_tokenizers/convert_tokenizer.py
@@ -114,7 +114,7 @@ def convert_tokenizer(
     can_use_sentencepiece = is_sentencepiece_model(tokenizer_object)
     is_unigram = can_use_sentencepiece and not is_sentencepiece_bpe_model(tokenizer_object)
     if isinstance(tokenizer_object, PreTrainedTokenizerBase):
-        if can_use_sentencepiece and (is_unigram or not tokenizer_object.is_fast or use_sentencepiece_backend):
+        if can_use_sentencepiece and (is_unigram or not tokenizer_object.is_fast or params.use_sentencepiece_backend):
             logger.info("Convert tokenizer using SentencePiece .model file.")
             ov_tokenizers = convert_sentencepiece_model_tokenizer(tokenizer_object, params)
         elif is_tiktoken_model(tokenizer_object):
--- a/python/openvino_tokenizers/convert_tokenizer.py
+++ b/python/openvino_tokenizers/convert_tokenizer.py
@@ -52,15 +52,13 @@ def convert_tokenizer(
         If provided, the `TokenzierConversionParams` object containing conversion parameters.
         If not provided, the parameters will be constructed from the other keyword arguments.
     """
-    (
-        +TokenzierConversionParams.__doc__
-        + """
+    + TokenzierConversionParams.__doc__
+    + """
     Returns:
     --------
     Union[Model, Tuple[Model, Model]]
         The converted tokenizer model, or a tuple tokenizer and detokenizer depending on with_detokenizer value.
     """
-    )
 
     if params is None:
         params = TokenzierConversionParams(
--- a/python/openvino_tokenizers/convert_tokenizer.py
+++ b/python/openvino_tokenizers/convert_tokenizer.py
@@ -10,77 +10,85 @@ from openvino.runtime import Model, Type
 from openvino.runtime.exceptions import OVTypeError
 
 from openvino_tokenizers.constants import UTF8ReplaceMode
-from openvino_tokenizers.utils import change_inputs_type, change_outputs_type, update_rt_info, TokenzierConversionParams
+from openvino_tokenizers.utils import (
+    change_inputs_type,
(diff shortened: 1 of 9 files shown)


python/openvino_tokenizers/utils.py
```diff
@@ -10,15 +10,66 @@
 from openvino import Model, Type
 from openvino.preprocess import PrePostProcessor
 from openvino.runtime import opset12 as opset
+from dataclasses import dataclass, field, asdict
+from transformers import PreTrainedTokenizerBase
```
apaniukov, line 14: I would avoid global import from transformers since it is an optional dependency.

pavel-esir: done

python/openvino_tokenizers/convert_tokenizer.py
```diff
@@ -10,14 +10,47 @@
...
+                          "convert_tokenizer(tokenizer_object, with_detokenizr=True, add_special_tokens=True,...). " \
+                          "Or within 'params' argument, e.g. " \
+                          "convert_tokenzier(tokenizer_object, params={'with_detokenizr': True, 'add_special_tokens': True, ...})"
+                    raise ValueError(msg)
+        
+        return func(*argc, **kwargs)
+    
+    # Embed convert_tokenizer docstring with TokenzierConversionParams docstring.
+    pos = func.__doc__.find('    Returns:')
+    wrapper.__doc__ = func.__doc__[:pos] + TokenzierConversionParams.__doc__ + '\n' + func.__doc__[pos:]
```
apaniukov, line 45: https://docs.python.org/3/library/functools.html#functools.wraps

python/openvino_tokenizers/convert_tokenizer.py
```diff
@@ -10,14 +10,47 @@
...
+            for key in TokenzierConversionParams.__match_args__:
+                if kwargs[key] is not None:
+                    msg = "Cannot specify both 'params' and individual convert_tokenizer arguments simultaneously. " \
+                          "Please pass all conversion params either individually, e.g. " \
+                          "convert_tokenizer(tokenizer_object, with_detokenizr=True, add_special_tokens=True,...). " \
+                          "Or within 'params' argument, e.g. " \
+                          "convert_tokenzier(tokenizer_object, params={'with_detokenizr': True, 'add_special_tokens': True, ...})"
+                    raise ValueError(msg)
+        
+        return func(*argc, **kwargs)
```
apaniukov, line 41:
```suggestion
    def wrapper(*args, **kwargs):
        params = None
        if len(args) > 1 and args[1] != None:
            params = args[1]
        if 'params' in kwargs:
            params = kwargs['params']
        
        if params is not None:
            for key in TokenzierConversionParams.__match_args__:
                if kwargs[key] is not None:
                    msg = "Cannot specify both 'params' and individual convert_tokenizer arguments simultaneously. " \
                          "Please pass all conversion params either individually, e.g. " \
                          "convert_tokenizer(tokenizer_object, with_detokenizr=True, add_special_tokens=True,...). " \
                          "Or within 'params' argument, e.g. " \
                          "convert_tokenzier(tokenizer_object, params={'with_detokenizr': True, 'add_special_tokens': True, ...})"
                    raise ValueError(msg)
        
        return func(*args, **kwargs)
```

python/openvino_tokenizers/convert_tokenizer.py
pavel-esir: Applied here @wraps and other stuff we discussed on teams.

python/openvino_tokenizers/convert_tokenizer.py
```diff
@@ -10,14 +10,55 @@
...
+    change_outputs_type,
+    update_rt_info,
+    TokenzierConversionParams,
+)
 
 logger = logging.getLogger(__name__)
 
 
+def capture_arg(func):
+    from functools import wraps
```
apaniukov, line 24: Move imports from the standard library to the top.

tests/tokenizers_test.py
```diff
@@ -894,3 +894,54 @@ def test_rt_info_sentencepiece(hf_sentencepiece_tokenizers, is_sentencepiece_bac
...
+]
+@pytest.fixture(scope="session", params=models_to_check_rt_info)
+def tokenizer_to_check_rt_info(request):
+    return get_hf_tokenizer(request, trust_remote_code=True)
+
+
+def test_rt_info_conversion_params(tokenizer_to_check_rt_info):
+    import openvino as ov
+    from openvino_tokenizers import convert_tokenizer
+    from openvino_tokenizers.utils import TokenzierConversionParams
```
apaniukov, line 913: These imports can be moved to the top.

apaniukov (approved):

tests/tokenizers_test.py
pavel-esir: done

python/openvino_tokenizers/convert_tokenizer.py
pavel-esir: done

PR #289, opened 2024-10-14
[GHA] Request wheels from OV provider explicitly

akladiev:
Since we didn't agree on downloading them by default.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -40,7 +40,7 @@ jobs:
       id: openvino_download
       with:
         platform: 'ubuntu22'
-        ov_artifact_name: 'wheels'
+        commit_packages_to_provide: 'wheels'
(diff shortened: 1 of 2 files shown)


ilya-lavrenov (approved):

.github/workflows/linux.yml
```diff
@@ -40,6 +40,7 @@ jobs:
       id: openvino_download
       with:
         platform: 'ubuntu22'
+        commit_packages_to_provide: 'wheels'
```
mryzhov, line 43: I though that the right argument is "ov_artifact_name: 'openvino_wheels'", why whees? How user should know the right name of other artifacts?

akladiev: It's the name of the folder stored on shared drive, it's seen in OV workflow, plus you've added this folder by yourself :)

PR #292, opened 2024-10-18
add packages versions to rt_info

pavel-esir:
- Added storing tiktoken, tokenizers, etc. version to rt_info as well.

image

CVS-146675


python/openvino_tokenizers/convert_tokenizer.py
```diff
@@ -71,7 +72,7 @@ def convert_tokenizer(
     use_sentencepiece_backend: bool = False,
     utf8_replace_mode: Optional[UTF8ReplaceMode] = None,
 ) -> Union[Model, Tuple[Model, Model]]:
-    """
+    """u
```
apaniukov, line 75:
```suggestion
    """
```

python/openvino_tokenizers/utils.py
```diff
@@ -222,12 +222,48 @@ def get_hf_tokenizer_attribute(
...
+    :param ov_tokenizer: Thes OpenVINO tokenizer model to update.
+    :type ov_tokenizer: openvino.Model
+
+    """
+    def get_package_version(name: str) -> str:
+        import importlib.metadata as metadata
+        try:
+            return metadata.version(name)
+        except metadata.PackageNotFoundError:
+            return None
```
apaniukov, line 239: No need for a nested local function here, it can be declared outside.

python/openvino_tokenizers/utils.py
```diff
@@ -222,12 +222,48 @@ def get_hf_tokenizer_attribute(
 ) -> Any:
     return next((value for attr in attributes if (value := getattr(hf_tokenizer, attr, None)) is not None), None)
 
+def update_rt_info_with_versions(ov_tokenizer: Model) -> None:
```
apaniukov, line 226:
```suggestion
def update_rt_info_with_environment(ov_tokenizer: Model) -> None:
```

pavel-esir: done

python/openvino_tokenizers/utils.py
pavel-esir: done

python/openvino_tokenizers/utils.py
```diff
@@ -224,11 +224,42 @@ def get_hf_tokenizer_attribute(
...
+    import importlib.metadata as metadata
+
+    try:
+        return metadata.version(name)
+    except metadata.PackageNotFoundError:
+        return None
+
+
+def update_rt_info_with_environment(ov_tokenizer: Model) -> None:
+    """Updates the runtime information with the versions of the HF tokenizer, Tiktoken, Sentencepiece and OpenVINO runtime.
```
apaniukov, line 237:
```suggestion
    """Adds package versions used for conversion to rt_info
```

PR #293, opened 2024-10-18
Revert "hotfix for RE2 segfaults"

rkazants:
Reverts openvinotoolkit/openvino_tokenizers#284

Breaks TensorFlow: https://github.com/openvinotoolkit/openvino/actions/runs/11400093905/job/31722751914


pavel-esir (approved):

pavel-esir:
Problem is here https://github.com/openvinotoolkit/openvino_tokenizers/blob/ce9d51c8d12fd27d24eb0f80b0226fd1f72ff5be/src/utils.cpp#L294

we didn't allocate enough memory. It's easy to fix. Will open another PR shortly.

PR #300, opened 2024-10-23
Fix RegexSplit Segfault

apaniukov:


benchmark/benchmark.py
```diff
@@ -48,6 +50,7 @@ def callback(
         end = perf_counter()
         times, start, idx = user_data
         times[idx] = end - start
+        # times[idx] = ir.latency * 1e-3
```
pavel-esir, line 53: probably left by mistake?

pavel-esir (approved):

benchmark/benchmark.py
apaniukov: Yes. Deleted.

PR #303, opened 2024-10-24
fix evaluate in concurrent scenario

pavel-esir:
Just added mutex to critical block where internal values of BPENode are changed.

CVS-152028


apaniukov (approved):

apaniukov (approved):

apaniukov (approved):

pavel-esir:
Checked OVMS locally in concurrent scenario and finally issue is gone

dkalinowski (approved):

PR #314, opened 2024-11-11
Update Regex For Clean Tokenization Spaces

apaniukov:
CVS-156773


src/regex_normalization.cpp
```diff
@@ -18,8 +18,20 @@ namespace {
...
+}
 
+/**
+ * @brief Fix old search pattern for backward compatibility
+ *
+ * @param search_pattern Search pattern to replace
+ * @return std::string Replaced search pattern
+ */
+std::string fix_search_pattern(std::string search_pattern) {
+    if (search_pattern == R"( ([\\.\\?\\!,])| ('[ms])| (') | ('[rv]e)| (n't))") {
```
pavel-esir, line 31: if there is pattern `r" ([\.\?\!\,])| ('[ms])| (') | ('[rv]e)"` this replace would not work, is it expected?

pavel-esir (approved):
in general LGTM, left one question

src/regex_normalization.cpp
apaniukov: This regex is for detokenization only and is not used anywhere. So I just deleted it.

PR #317, opened 2024-11-14
Update Prepend Regex

apaniukov:
CVS-156583


src/regex_normalization.cpp
```diff
@@ -18,20 +17,36 @@ namespace {
  * @return std::string Reformatted replace pattern
  */
 std::string reformat_replace_pattern(std::string replace_pattern) {
-    return std::regex_replace(replace_pattern, std::regex(R"((?:\\)([0-9]+))"), R"($$1)");
+    for (char i = '1'; i <= '9'; ++i) {
```
pavel-esir, line 20: Why did you change this implementation?

pavel-esir (approved):

src/regex_normalization.cpp
apaniukov: Regex didn't work in some cases.

PR #322, opened 2024-11-19
Turn on UTF8Validate.REPLACE by default

pavel-esir:
for `TextCallbackStreamer` to handle unfinished characters correctly

https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/text_callback_streamer.cpp#L26-L27

CVS-157110


apaniukov (approved):

pavel-esir:
there is still a typo, please don't merge. I will update when all typos are fixed

pavel-esir:
fixed and checked. Both cli and python api correctly works by default and when 'ignore', 'replace', 'disable' are specified.

PR #334, opened 2024-12-04
Print debug errors only if ENV VAR is set

pavel-esir:


apaniukov (approved):

pavel-esir:
Need to port to master

PR #388, opened 2025-01-27
Fix RegexSplit

pavel-esir:
- If there are splits split_pattern should be in `input[5 + 1]`, corrected that.
- `RegexSplit` was incorrectly getting input pattern in `evaluate`. Issue was not visible before because the first compilation was done in `validate_and_infer_types` and `compile_pattern_if_necessary` did an early return. But when model is used with  `ov::cache_dir("./cache")` then `compile_pattern_if_necessary` required actual string value and issue becomes visible.

- Ticket: CVS-161231


pavel-esir:
rebuild

src/regex_split.cpp
```diff
@@ -163,13 +165,11 @@ bool RegexSplit::evaluate(ov::TensorVector& outputs, const ov::TensorVector& inp
     auto ends   = inputs[3].data<const int32_t>();
     auto chars  = inputs[4].data<const uint8_t>();
 
-    outputs[4] = inputs[4];
     const size_t num_rows = inputs[0].get_size();
     const size_t num_chars = inputs[4].get_size();
 
     bool * skips;
```
pavel-esir, line 171:
```suggestion
    bool* skips;
```

pavel-esir:
need to add unit-tests with reloading models with  `ov::cache_dir("./cache")`

apaniukov (approved):

andrei-kochin (changes requested):
No committee approval received yet

pavel-esir:
Opened PR into master https://github.com/openvinotoolkit/openvino_tokenizers/pull/390
Since priority of the bug is lowered, no need to port to release. Closing this PR

PR #431, opened 2025-03-10
Add Unigram Tokenizer Implementation

apaniukov:
- Add UnigramTokenizer operation
- Change the default conversion behaviour for Fast tokenizer to use new Unigram implementation instead of the Sentencepiece backend
- Add support for Strip normalization operation
- Separate Sentencepiece backend tests from our implementation of BPE and Unigram
- Fix shape estimation for RegexSplit

CVS-131053, CVS-163883


ilya-lavrenov:
Will it also fix https://github.com/openvinotoolkit/openvino.genai/pull/1774 ?
I remember, that FLUX does not work there

pavel-esir (approved):

apaniukov:
> Will it also fix [openvinotoolkit/openvino.genai#1774](https://github.com/openvinotoolkit/openvino.genai/pull/1774) ? I remember, that FLUX does not work there

Yes, this should fix it.

PR #450, opened 2025-03-21
Use const pointers for const tensors

praasz:
### Description
- Update pointers to tensor data to be constant when tensor is const.

### Blocking:
- openvinotoolkit/openvino#29594


apaniukov:
image

It seems like `skips` is not passed correctly.

apaniukov (approved):

PR #473, opened 2025-04-29
[DRAFT] UV dependency management

mryzhov:
Enabled uv dependency management


pyproject.toml
```diff
@@ -120,3 +102,33 @@ requires = [
...
+[tool.uv]
+package = false
+default-groups = ["dev"]
+
+[tool.uv.sources]
+openvino = { index = "openvino-nightly" }
+
+[[tool.uv.index]]
+name = "openvino-nightly"
+url = "https://storage.openvinotoolkit.org/simple/wheels/nightly"
```
ilya-lavrenov, line 133: is it possible to add pre-release as well ?

mryzhov: yes, we can define 2 source links here

mryzhov: tested, we could not set to sources for one package without marker condition, the same story for poetry

PR #496, opened 2025-05-26
[GHA] Enabled product manifest.yml

mryzhov:
- Internal CI uses manifest.yml to pick the build artifacts
- Added the dev suffix and build tag in the wheel name (aligning with ov)

depends on https://github.com/openvinotoolkit/openvino/pull/30742


.github/workflows/windows.yml
```diff
@@ -283,6 +284,17 @@ jobs:
         with:
           toolset: 14.40 # v2022
           
+      - name: Generate product manifest
+        id: create_manifest
```
akladiev, line 288: Can we create it right after repo clone now?

.github/workflows/mac.yml
```diff
@@ -20,6 +20,7 @@ env:
   CMAKE_CXX_COMPILER_LAUNCHER: ccache
   CMAKE_C_COMPILER_LAUNCHER: ccache
   CCACHE_MAXSIZE: 100Mi
+  BASE_PRODUCT_TYPE: private_darwin
```
mryzhov, line 23:
```suggestion
  BASE_PRODUCT_TYPE: public_darwin
```

akladiev (approved):

PR #542, opened 2025-08-08
Allow to specify value 2 in input to pad_right

pavel-esir:
- This allow to set value integer values (instead of bool) to `pad_right` input. This is necessary because if user called in ov::genai::Tokenizer` method `encode` with no padding value it should pad to the default value which is used during model conversion which is stored in the attribute, but since we don't have access to the attribute we specify by value `2` on `pad_right` input that op should rely on attribute value. See more details in the comment.
- Ticket CVS-163374


pavel-esir:
Managed to get default value of attribute inside GenAI transformation. So there is no need for one more value indicating take default value form the attr.

PR #636, opened 2026-03-27
ONNX Tokenizers Support: Label Encoder, StringNormalizer, Tokenizer, TFID Vectorizer. 

sfatimar:
[CVS-181364](https://jira.devtools.intel.com/browse/CVS-181364) : We want to enable StringNormalizer ops as part of a collection of ops we want to enable tokenizer onnx custom ops.

1. StringNormalizer 
This is implemented with the combination of CaseFold and RegexNormalization OV IR Ops.

CaseFold Op is extended to include UPPER case for empty encoding. Locale is not supported. Need to check with OpenVINO team how to implement this.ONNX Custom Ops Implementation is added in OpenVINO Tokenizer extension library, Reference https://onnx.ai/onnx/operators/onnx__StringNormalizer.html 
2. Label Encoder : This is implemented with Vocabulary Encoder. Currently we only support string tensor as input encoded to int64_t values . Reference https://onnx.ai/onnx/operators/onnx_aionnxml_LabelEncoder.html  
3. Tokenizer: This is implemented through CumSum, Regexsplit and RaggedToDense implementation for various attributes. Curretly impleementation does not support sep attribute . https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.Tokenizer
4. TFID Vectorizer. : This is implemented through one hot encoder and VocabEncoder . We only support  unigrams for present implementation and only TF mode is supported for unigrams. Reference https://onnx.ai/onnx/operators/onnx__TfIdfVectorizer.html implementation: Addition of other attributes is TBD. Current implementation supports Model


apaniukov (changes requested):

apaniukov (approved):

mryzhov:
build_jenkins

PR #660, opened 2026-06-03
Add ai.onnx.contrib string-op ONNX conversion extensions

mvafin:
### Details
- Adds ONNX Frontend conversion extensions for five ai.onnx.contrib operators emitted by onnxruntime_extensions, enabling read_model on ONNX models that embed SentencePiece tokenization and string post-processing (e.g. seq2seq pipelines exported with ORT extensions):

  - SentencepieceTokenizer — accepts both the 6-input form and the optional 7-input fairseq variant (validated to be false; fairseq-   style id remapping is unsupported and fails fast).
  - SentencepieceDecoder — maps token ids back to strings via SentencepieceDetokenizer (fairseq likewise validated false).
  - VectorToString — id→string lookup with unk fallback for out-of-range ids.
  - StringJoin — joins string elements along an axis with a separator.
  - StringSplit — splits strings by a delimiter into a sparse (indices, values, dense_shape) representation.
- All ops are registered as ConversionExtensions in ov_extension.cpp and resolve automatically once the tokenizers extension is loaded.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,13 +121,18 @@ optional = true
 
 [tool.poetry.group.dev.dependencies]
 ruff = ">=0.1.0,<=0.12.8"
-onnx = ">=1.13.0"
 bandit = ">=1.7.0,<=1.8.6"
 pytest = ">=7.0.0,<=8.3.5"
 pytest-xdist = ">=3.4.0,<=3.8.0"
 pytest-harvest = ">=1.6.0,<=1.10.0"
 pandas = ">=2.0.0,<=2.3.1"
 jinja2 = ">=3.0.0,<=3.1.6"
+
+[tool.poetry.group.onnx_tests]
+optional = true
+
+[tool.poetry.group.onnx_tests.dependencies]
+onnx = ">=1.13.0"
 
 [tool.poetry.group.benchmark]
 optional = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,6 +133,7 @@ optional = true
 
 [tool.poetry.group.onnx_tests.dependencies]
 onnx = ">=1.13.0"
+sentencepiece = ">=0.1.96"
 
 [tool.poetry.group.benchmark]
 optional = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,6 +121,7 @@ optional = true
 
 [tool.poetry.group.dev.dependencies]
 ruff = ">=0.1.0,<=0.12.8"
+onnx = ">=1.13.0"
 bandit = ">=1.7.0,<=1.8.6"
 pytest = ">=7.0.0,<=8.3.5"
 pytest-xdist = ">=3.4.0,<=3.8.0"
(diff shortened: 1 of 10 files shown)


src/contrib_string_ops.hpp
```diff
@@ -0,0 +1,90 @@
...
+// ai.onnx.contrib.StringJoin equivalent. Joins string elements of a decomposed
+// (begins, ends, chars) string tensor along `axis` using a scalar `sep`.
+//
+// Inputs (in order):
+//   0: input begins   (i32, tensor shape S)
+//   1: input ends     (i32, tensor shape S)
+//   2: input chars    (u8,  1D)
+//   3: sep begins     (i32, 0D/1D, single element)
+//   4: sep ends       (i32, 0D/1D, single element)
+//   5: sep chars      (u8,  1D)
```
apaniukov: We can use one input when single element string is expected, for example in RegexSplit: https://github.com/openvinotoolkit/openvino_tokenizers/blob/1f0085d8c2f4846ade923c8eb0a96a460fdd239c/src/regex_split.cpp#L145

Same for `delim` in `ContribStringSplit`.

src/contrib_string_ops.cpp
```diff
@@ -0,0 +1,335 @@
...
+    size_t out_size = ov::shape_size(out_shape);
+    size_t axis_size = in_shape[axis];
+
+    // Output strides (over out_shape) for unraveling output linear index.
+    std::vector<size_t> out_strides(out_shape.size(), 1);
+    for (int i = static_cast<int>(out_shape.size()) - 2; i >= 0; --i) {
+        out_strides[i] = out_strides[i + 1] * out_shape[i + 1];
+    }
+
+    std::vector<std::string> result(out_size);
```
apaniukov: Could we avoid intermediate result buffer and write directly to `outputs`? `total_chars` can be computed beforehand for `outputs[2].set_shape`.

pyproject.toml
```diff
@@ -121,6 +121,7 @@ optional = true
 
 [tool.poetry.group.dev.dependencies]
 ruff = ">=0.1.0,<=0.12.8"
+onnx = ">=1.13.0"
```
apaniukov: Could you move the ONNX to a separate dependency group?

src/contrib_string_ops.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+    // Pass 1: compute total output chars to pre-size the output buffer.
+    size_t total_chars = 0;
+    for (size_t oi = 0; oi < out_size; ++oi) {
+        size_t base = compute_base_offset(oi);
+        for (size_t a = 0; a < axis_size; ++a) {
+            size_t idx = base + a * in_strides[axis];
+            if (a > 0) total_chars += sep.size();
+            total_chars += static_cast<size_t>(ends[idx] - begins[idx]);
+        }
```
apaniukov, line 171: Aren't it just something like `total_chars += ends[base + axis_size] - begins[base] + (axis_size-1) * sep.size()`?

mvafin: This is only true for the innermost axis (stride == 1) with contiguous string storage.

apaniukov: We can instantiate `total_chars` with `out_size * (axis_size - 1) * sep.size()` and delete `if` from the inner loop entirely.
Also, `in_strides[axis]` doesn't change, so maybe it is better to instantiate it outside the loop as well.

mvafin: done

apaniukov (approved):

PR #676, opened 2026-06-29
Use `uv` in CI 

apaniukov:


.github/actions/install_uv/action.yml
```diff
@@ -0,0 +1,13 @@
+name: Install uv
```
mryzhov: We don't need to introduce the custom action, just use the astral-sh directly in the pipleines

apaniukov: Done

mryzhov (approved):

PR #677, opened 2026-06-30
[EXPERIMENT][WIP] fix(VocabDecoderStep): avoid zero-dim Const when skip_special_tokens=False

mlukasze:
> ⚠️ AUTOMATICALLY GENERATED BY OMEGA AGENT — REQUIRES HUMAN REVIEW ⚠️
> This PR was created by an AI agent as part of automated model enablement.
> A human maintainer must review and approve it before it can be considered for merge.
> Do **NOT** merge without human review and sign-off.

## Summary

When converting a detokenizer with `skip_special_tokens=False`, `VocabDecoderStep.get_ov_subgraph()` builds a `Slice` with `stop=0`, which constant-folds to a **zero-dimensional `Const`** wired into `VocabDecoder` port-4 (the skip-tokens list). When such a detokenizer is loaded together with a stateful model that has KV-cache `ReadValue`/`Assign` nodes (e.g. via `openvino_genai.VLMPipeline`/`LLMPipeline`), the Intel CPU plugin crashes with:

```
Check 'edge' failed at src/core/src/node.cpp:712: Node ReadValueXXXX contains dead weak ptr
```

## Fix

When `do_skip_tokens=False`, omit port-4 entirely and use the 4-input `VocabDecoder` form (the op already treats a missing skip-tokens input as "skip nothing" via the empty `m_skip_tokens` attribute), instead of creating a zero-dim `Const`. This eliminates the zero-dim constant that triggers the plugin crash, with no behavioural change (skip_special_tokens=False already means "skip nothing").

## Test

Adds a regression test asserting that a detokenizer converted with `skip_special_tokens=False` produces no zero-dimensional `Const` feeding the `VocabDecoder`, and round-trips special tokens correctly.

## Context

Found while enabling `opendatalab/MinerU2.5-Pro-2604-1.2B` (a Qwen2-VL document-parsing VLM) for OpenVINO. That model relies on custom structural special tokens (`<fcel>`, `<ched>`, `<|md_start|>`, …) in its generated Markdown; `skip_special_tokens=False` is required to preserve them in `VLMPipeline` output.

## Follow-up

A separate hardening could be made in the OpenVINO CPU plugin to tolerate zero-dim `Const` inputs to custom ops gracefully rather than crashing; tracked as a follow-up issue. This PR fixes the issue at the tokenizer-graph layer, which is the correct place.

### DIFF
--- a/python/openvino_tokenizers/tokenizer_pipeline.py
+++ b/python/openvino_tokenizers/tokenizer_pipeline.py
@@ -1301,15 +1301,25 @@ class VocabDecoderStep(DecodingStep):
             vocab_outputs = create_string_constant_node(self.vocab)
         input_nodes.extend(vocab_outputs)
 
-        # Put constant with skip tokens even if do_skip_tokens=False, so that it can be switched on/off at runtime.
-        # Slice through all skip tokens if flag is true, else slice to get an empty tensor.
-        stop_const = op.Constant(Type.i32, Shape([1]), [np.iinfo(np.int32).max if self.do_skip_tokens else 0])
-
-        zero_const = op.Constant(Type.i32, Shape([1]), [0])
-        one_const = op.Constant(Type.i32, Shape([1]), [1])
-        skip_tokens_const = op.Constant(Type.i32, Shape([len(self.skip_tokens)]), self.skip_tokens)
-        sliced_skips = opset.slice(skip_tokens_const, zero_const, stop_const, one_const).outputs()
-        input_nodes.extend(sliced_skips)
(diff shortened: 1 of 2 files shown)


apaniukov:
`skip_special_tokens` parameter can be dynamically adjusted after export in GenAI: https://github.com/openvinotoolkit/openvino.genai/blob/61ed04788e5b7476d5c877f928fae90d245973c7/tests/python_tests/test_tokenizer.py#L353

Deleting special tokens constant from the graph will break this functionality. 
PR for the CPU plugin is better solution.

PR #696, opened 2026-07-14
Fix missing Abseil symbols on Android after OpenVINO Protobuf update

bumbosiepsak:
### Details:
- Added missing unconditional dependency of `openvino_tokenizers` target on `absl`
- Kept explicit dependency of `charsmap generator` on `absl` (despite inheritance from `sentencepiece`) as a safety net in case of more sophisticated installation patterns. The required targets list was taken from `sentencepiece` code.
- This proved necessary after a `Protobuf` update in `OpenVINO` (to `4.25.1`)

This is a second attempt at making OpenVINO build with an updated Protobuf

### Tickets:
 - CVS-187110

### AI Assistance:
 - *AI assistance used: yes*

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -67,12 +67,6 @@ endif()
 # Dependencies
 #
 
-set(ABSL_DEPENDENCIES
-    absl::flat_hash_set
-    absl::raw_hash_set
-    absl::string_view
-)
-
 include(FetchContent)
 
 # This option must be defined early because it affects SentencePiece configuration
@@ -149,7 +143,6 @@ function(ov_tokenizers_link_sentencepiece TARGET_NAME)
     else()
       target_link_libraries(${TARGET_NAME} PRIVATE sentencepiece)
     endif()
-    target_link_libraries(${TARGET_NAME} PRIVATE ${ABSL_DEPENDENCIES})
   else()
     if(SPM_PROTOBUF_PROVIDER STREQUAL "internal")
       if(SPM_ABSL_PROVIDER STREQUAL "package")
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -253,7 +246,10 @@ if(REGENERATE_PRECOMPILED_CHARSMAP)
         target_link_libraries(${CHARSMAP_GENERATOR_NAME} PRIVATE ${sp_target})
       endif()
     endforeach()
-    target_link_libraries(${CHARSMAP_GENERATOR_NAME} PRIVATE ${ABSL_DEPENDENCIES})
+    target_link_libraries(${CHARSMAP_GENERATOR_NAME} PRIVATE
```
bumbosiepsak, line 249: `absl` dependencies should be inherited from `sentencepiece`. However this works only for in-source builds. In case current configuration doesn't offer `SentencePieceConfig.cmake`, an explicit dependency declaration is needed.

I restored the original code (prior to my changes last week) and revised the list of `absl` sub-libs required by `sentencepiece`.

src/CMakeLists.txt
```diff
@@ -407,7 +403,13 @@ if(LINUX)
 endif()
 
 target_compile_definitions(${TARGET_NAME} PRIVATE IMPLEMENT_OPENVINO_EXTENSION_API)
-target_link_libraries(${TARGET_NAME} PRIVATE openvino::runtime openvino::threading)
+target_link_libraries(${TARGET_NAME} PRIVATE
+    absl::flat_hash_map
```
bumbosiepsak: The `openvino_tokenizers` target unconditionally needs these `absl` libraries due to explicit usage:
| File | absl usage |
|---|---|
| `src/utils.hpp` | `absl::string_view` in `PCRE2Wrapper` constructor/method signatures |
| `src/utils.cpp` | `absl::string_view` in `PCRE2Wrapper::PCRE2Wrapper()` and `PCRE2Wrapper::substitute()` |
| `src/sentence_piece.hpp` | `absl::flat_hash_map<std::string, int32_t>` in `SentencepieceTokenizer` private members (`absl/strings/str_format.h`, `absl/container/flat_hash_map.h`) |
| `src/sentence_piece.cpp` | `absl::string_view` as encode function argument type |
| `src/bpe_tokenizer.hpp` | `absl::flat_hash_map<...>` as `Merges` type alias (`absl/container/flat_hash_map.h`) |
| `src/bpe_tokenizer.cpp` | `absl::StrFormat` for byte fallback token formatting (`absl::StrFormat("<0x%02X>", ...)`) |
| `src/vocab_encoder.cpp` | `absl::flat_hash_map<std::string, T>` for vocab storage |
| `src/unigram_tokenizer.hpp` | `absl::string_view` in `UnigramTokenizerImpl::tokenize()` signature |
| `src/unigram_tokenizer.cpp` | `absl::string_view` passed to `tokenize()` |

apaniukov (approved):

PR #714, opened 2026-08-04
build: expose SentencePiece generated headers to tokenizer targets

stevegrubb:
When OpenVINO Tokenizers builds SentencePiece from source, SentencePiece generates config.h in its binary directory. Targets linked through ov_tokenizers_link_sentencepiece do not include that directory and can fail to find the generated header.

Add sentencepiece_BINARY_DIR as a private system include for the source-built SentencePiece path. This leaves preexisting package targets and the installed OpenVINO Tokenizers interface unchanged.


apaniukov:
build_jenkins

apaniukov (approved):

PR #731, opened 2026-08-24
Forward checker options to GenAI Tokenizer.encode

MaxxxDong:
#### Description

`check_tokenizer.py` builds Hugging Face reference results with runtime `truncation`, `max_length`, max-padding, and padding-side options, but its `openvino_genai.Tokenizer.encode` calls did not forward the same options on the single, batch, and paired-input paths. This can make the checker report mismatches even when both tokenizers support the same runtime configuration.

This change:

- forwards the same encode options to `openvino_genai.Tokenizer` for single, batch, and paired inputs;
- preserves `truncation=False` unless the CLI or the built-in fixed-length batch case enables it;
- applies the CLI `max_length` when max padding is requested, even without truncation;
- adds a focused regression test for the shared GenAI encode parameters.

#### Reproduction

Environment used for the original reproduction:

- Windows, Python 3.12.10
- OpenVINO 2026.3.0
- OpenVINO Tokenizers 2026.3.0.0
- OpenVINO GenAI 2026.3.0.0
- Transformers 5.15.1
- `Qwen/Qwen3.5-4B` tokenizer revision `851bf6e806efd8d0a36b00ddf55e13ccb7b8cd0a`

With the unmodified checker, the built-in `batch/max_length=32` case reports a padding mismatch. Explicit truncation/max-length and max-padding combinations additionally expose single and paired-input mismatches because the Hugging Face side receives runtime options that the GenAI side does not.

The same tokenizer snapshot was checked with these option sets:

| Case | max_length | truncation | max padding | Baseline Step 4/5 | Patched Step 4/5 |
| --- | ---: | --- | --- | --- | --- |
| A | None | false | false | Step 5 batch mismatch | pass |
| B | 64 | false | false | Step 5 batch mismatch | pass |
| C | 64 | true | false | single + batch mismatch | pass |
| D | 64 | false | true | single + batch + pair mismatch | pass |
| E | 64 | true | true | single + batch + pair mismatch | pass |

The current `master` Python checker requires the matching 2026.4 development OpenVINO stack for a complete conversion run. The Windows matrix therefore used the same pre-converted 2026.3 tokenizer artifact for baseline and patched Step 3/4/5 comparisons; no installed package was modified. Repository CI remains the authoritative full-build check against current `master`.

#### Root cause

The Hugging Face calls and GenAI calls were configured asymmetrically:

- the single-input GenAI call only received `add_special_tokens`;
- the batch path omitted effective truncation and max-padding options;
- the paired-input GenAI call received no runtime encode options.

#### Tests

- `ruff check python/openvino_tokenizers/cli_tools/check_tokenizer.py tests/check_tokenizer_test.py`
- `pytest -q tests/check_tokenizer_test.py`
- Windows Qwen tokenizer A-E matrix described above

#### Compatibility and scope

This only changes the checker. It does not change tokenizer conversion or runtime defaults. In particular, it retains the current `truncation=False` default and only mirrors the options already used for the Hugging Face reference call.


python/openvino_tokenizers/cli_tools/check_tokenizer.py
```diff
@@ -351,7 +370,10 @@ def step_test_genai(
                 truncation=truncation,
                 max_length=max_length,
             )["input_ids"][0]
-            genai_ids = genai_tok.encode(test_str, add_special_tokens=add_spec).input_ids.data[0]
+            genai_ids = genai_tok.encode(
+                test_str,
+                **_genai_encode_params(add_spec, use_max_padding, max_length, None, truncation),
```
apaniukov: Better to form this parameters outside `for test_str in ALL_TEST_STRINGS:` loop. Also, the function is less readable then dict creation, since it contains all parameter names. Something like:
```python
genai_tok.encode(test_str, **(ov_params | {"add_special_tokens": add_spec})).input_ids.data[0]
```

MaxxxDong: Updated: moved the common parameters outside the string loop and replaced the helper with explicit dictionaries.

MaxxxDong:
Hi @apaniukov, I’ve addressed your review feedback in 0b6a89b and replied to the review thread. When convenient, could you please take another look and approve the pending workflow runs? The branch is currently behind but cleanly mergeable; I’m happy to update it if you prefer. Thanks!

apaniukov (approved):

apaniukov:
build_jenkins

apaniukov:
build_jenkins

MaxxxDong:
Hi @apaniukov @akladiev, PR #731 was removed from the merge queue because the merge-group jobs could not resolve dependencies after the automated 2026.5 version bump in #741. The same master SHA `824033c` is currently failing its Windows, Linux, macOS, and Manylinux push workflows.

`uv` reports that `openvino>=2026.5.0.dev0,<2026.6.dev0` is unavailable for the Python >=3.13 resolution split. PR #731 itself had all GitHub Actions and Jenkins checks passing before entering the queue.

Could you please take a look at the base dependency issue and requeue #731 once master is green? Thanks!

apaniukov:
build_jenkins

PR #735, opened 2026-08-26
cmake: add an opt-in system PCRE2 mode

stevegrubb:
OpenVINO Tokenizers always declares and populates its pinned PCRE2 source through FetchContent. This prevents network-free system-package builds and can put a second PCRE2 implementation into a process that already uses the system library.

Add a default-off ENABLE_SYSTEM_PCRE2 option. When enabled, use pkg-config's libpcre2-8 module and link its imported target. Report actionable errors when pkg-config or the PCRE2 development metadata is unavailable. Define the target at project scope so every Tokenizers subdirectory can use it without exposing a GLOBAL target to embedding projects. With the option off, retain the existing pinned FetchContent behavior. Both paths continue to use 8-bit PCRE2 code units.


CMakeLists.txt
```diff
@@ -97,6 +98,40 @@ if(BUILD_CPP_EXTENSION)
...
+        if(NOT PkgConfig_FOUND)
+            message(FATAL_ERROR
+              "ENABLE_SYSTEM_PCRE2=ON requires pkg-config to locate libpcre2-8. "
+              "Install pkg-config and the PCRE2 development files, or configure "
+              "with ENABLE_SYSTEM_PCRE2=OFF to use the bundled PCRE2 source.")
+        endif()
+
+        # Define the imported target here so every Tokenizers subdirectory can
+        # use it without leaking a GLOBAL target into an embedding project.
+        pkg_check_modules(PCRE2 QUIET IMPORTED_TARGET libpcre2-8)
```
apaniukov, line 112: Could we use a project-specific pkg-config prefix here, such as `OV_TOKENIZERS_PCRE2`? When this project is included through `add_subdirectory()`, a parent may already define the generic `PkgConfig::PCRE2` target.

apaniukov:
build_jenkins

PR #742, opened 2026-08-28
Reimplement Special Tokens Split

apaniukov:


pavel-esir (approved):
in general looks good to me

